PR #8, opened 2017-11-23
content_for page titles

andycroll:
Trying to get ahead.


_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+layout: article
+category: ruby
+image:
+  base: '2017/set-title-in-the-view-template-using-content-for'
+  alt: 'Books'
+  credit: 'Annie Spratt'
+  source: 'https://unsplash.com/photos/-9vMBjrU-RA'
+---
+
+Rails has `content_for` as it’s primary way to store content in one place for use in other views, layouts or helpers. See `content_for` in the Rails documentation at [in the ActionView helpers](http://api.rubyonrails.org/v5.1.3/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for).
```
nodunayo: * Rails _uses_ `content_for` as _its_ primary way to store content...
(Emphasis for you, not for the post. :-) )

nodunayo: *Find out more about `content_for` in the Rails documentation in the [ActionView helpers] section

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Instead of…
+
+...using an instance variable in the controller.
```
nodunayo: I would replace the full-stop at the end of this line with a colon.

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+In the `<head>`.
+
+```erb
+<title><%= content_for?(:html_title) ? yield(:page_title) : "Default" %></title>
+```
+
+
+## But why?
+
+Using `content_for` in this way keeps information about how to render the page inside the relevant view template. It also avoids cluttering the controller with non-business logic.
```
nodunayo: Take out 'also' in the second sentence here, i.e. "It avoids cluttering..." This is because I think this is one main point, i.e: this is the right place for it to go. Or you can make it one long sentence: "...inside the relevant view template and it avoids cluttering the controller..." Up to you!

You also have an 'also' in the next paragraph..so taking it out here helps it to flow better.

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+
+You also avoid creating yet another 'magic' instance variable that gets automatically revealed to the view by Rails.
```
nodunayo: I'm tempted to shorten this sentence, stopping it after 'variable' and then add it on to the paragraph before...

🤔

The latter half of it ("...that gets automatically revealed...") sounds a bit clunky. Maybe you can reword what you're trying to say instead?

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+
+Another heuristic for using this technique is that when rendering JSON and/or AJAX from the same restful action, those versions don't need a page title, indicating it belongs in the HTML view template.
```
nodunayo: *RESTful? (I dunno what the convention is..maybe you're right!)

nodunayo: I think what you're trying to say is:

> Here's another way you know I'm right about how you should set your page titles. Rending JSON or AJAX doesn't require adding a page title. Therefore, it's clear that the info should live with the relevant view.

Is that correct?

If so, I think you can reword this paragraph to be more straightforward.

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,71 @@
...
+
+You _can_ also use the `content_for` in multiple places in your views and the values are concatenated when they are `yield`ed in the layout. Not so useful in this case... but good to know.
```
nodunayo: Not sure what you mean by "these values are concatenated...". What do you mean? 

Just for me can you give me an example so I can think of a better rewording. Or maybe it's fine once I understand what you mean!

nodunayo (dismissed):

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
nodunayo: Emphasis on 'I'.

(Especially as it's not a complete sentence. You're trying to say: "Instead of X, use Y." Not: "Instead of X. Use Y." But...maybe you are! :P )

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
andycroll: If you use

```erb
<% content_for(:thing) { 'Nadia' } %>
<% content_for(:thing) { 'is ace' } %>
```

When you `yield(:thing)` you get `'Nadia is ace'`.

Think I might remove this it's tangential at best.

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
andycroll: I weep for my own grammar. It's super-weird to me that this doesn't stand out reading back my own writing, but I spot it so easily in other people's stuff.

nodunayo: Oh yeah, always easier when you're reading somebody else's stuff.

I like to think I'm very good at this but JB always manages to make MASSIVE improvements to anything I write, so...

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
nodunayo: Got it!

Thanks for the example. :-)

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,69 @@
...
+In the `<head>`.
+
+```erb
+<title><%= content_for?(:html_title) ? yield(:page_title) : "Default" %></title>
+```
+
+
+## But why?
+
+Using `content_for` keeps information about how to render the page inside the relevant view template, it avoids cluttering the controller with non-business logic.
```
nodunayo: Full-stop after 'template', rather than a comma.

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
```diff
@@ -0,0 +1,69 @@
...
+
+You avoid creating yet another 'magic' instance variable that gets automatically passed to the view by Rails.
+
+Here's another way you know I'm right about how you should set your page titles. Rendering JSON or AJAX from the same RESTful controller action doesn't require adding a page title. Therefore, it's clear that this information should live with the relevant HTML view template.
```
nodunayo: Colon after 'titles' rather than a full-stop.

nodunayo: I get what you're saying...but I'm still not convinced that it is 'clear'...I wonder if there's another way you can say this...

We need something that explains either what it would mean if you did have to set a page title there...or why it's key that you don't have to set a page title there...

Like, I may be reading this sentence — Rendering JSON or AJAX from the same RESTful controller action doesn't require adding a page title. — not understand the significance of that statement so think: 🤷🏾‍♀️...then the next thing you say is: "therefore it's clear..." and I'm thinking: no, it's not..

What the "Rendering JSON or AJAX.." sentence says to me is that it's clear where the info shouldn't go/doesn't need to go...but not clear that it should therefore be in the view.

Do you perhaps want to get across: therefore, Rails is built with the expectation that page title info is in the view hence why you don't need to set one when rendering JSON or AJAX...?

Which bit proves that it still can't live in the controller?

This has been one long ramble...but hopefully you get what I'm saying.

nodunayo (changes requested):

nodunayo:
It's just that one paragraph, Andy...!!

_posts/2017/2017-12-11-set-title-in-the-view-template-using-content-for.md
andycroll: How about...

Not convinced? If you were to render a JSON resource from the same controller action, using Rails' `respond_to` [syntax](http://guides.rubyonrails.org/action_controller_overview.html#rendering-xml-and-json-data), the resulting JSON response wouldn't need the content for the HTML `<title>`. Thus the information about setting the HTML `<title>` belongs _in_ the HTML template.

nodunayo (commented):
I actually prefer the old 'Why not?'..but hey ho! :P

nodunayo (approved):

PR #9, opened 2017-12-11
Avoid a double negative

andycroll:
Had a more complex version of this at one point but realised I was covering too much ground.


_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+layout: article
+category: ruby
+image:
+  base: '2017/avoid-double-negatives'
+  alt: 'Two directions sign'
+  source: 'https://unsplash.com/photos/lPQIndZz8Mo'
+  credit: 'Pablo Garcia Saldaña'
+---
+
+Ruby has a bunch of delightful syntactic sugar. One example of this is `unless`, which you can use in place of an `if` with a negative.
```
nodunayo: maybe at "..built in" at the end of the this first sentence?

nodunayo: *in place of using `if` combined with a negative statement.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+However given the flexibility of Ruby's syntax it is easy to make code harder to understand than it needs to be.
```
nodunayo: Add a comma after "However" and another after "syntax".

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Instead of…
+
+...overcomplicating your conditionals.
```
nodunayo: I'd add a colon instead of a full-stop here...

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+# Instead of Examples 1,2 & 3
+if something
+  # do something
+end
+```
+
+
+## But why?
+
+Ruby’s conditional syntax is ‘truthy’. In Ruby’s case that means `nil` is equivalent to `false`. There is no need to do a negated `#nil?` check on an object as it is redundant as shown in the first two examples.
```
nodunayo: Get rid of "In Ruby's case" and just start at "That means..."
Because we know we're in the context of Ruby.

nodunayo: Change the last sentence here to read: Therefore, a negated `#nil?` check on an object is redundant.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+The syntax of `!!` in the third example converts a value to boolean, but it is unnecessary and likely obscures your intention. If you _really_ want to do a `nil` check, you should use `#nil?` instead.
```
nodunayo: What is the intention here?

Maybe start with explaining why you'd want to use `!!`, i.e. why you want a boolean...then it'll be easier to show why just using `#nil?` makes sense.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+You should definitely avoid using an `else` block with an `unless`: the `else` block is a double negative of the condition. Instead, switch the `unless` for an logically equivalent `if` and swap the code in the successful conditional block for the code in the `else` block. This might require some refactoring to the condition to make the resulting code easy to understand.
```
nodunayo: This paragraph is hard to follow.

Can you use an code example (just for me) to illustrate what you're attempting to say?

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+It is also _much_ harder to reason when using an `unless` conditional with any boolean algebra (`&&` or `||`). It is an easy way to confuse yourself! The conditional `unless one && two` is the same code as `if !one || !two`. In this case the complex `if` condition is an opportunity to refactor the conditional into a well named method so it can be better understood.
```
nodunayo: *The conditional `unless one && two` _is equivalent to_ `if !one || !two`.

Given that it's not the same code...if you know what I mean...

nodunayo: *well-named

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## Why not?
+
+I’m definitely not saying don't use `unless`, but it's important to remember it is _sugar_ and thus it is important only to use it where it makes the resulting code clearer.
```
nodunayo: Great final section! :_)

nodunayo (dismissed):

_posts/2017/2017-12-31-avoid-double-negatives.md
andycroll: Reasoning about 

```ruby
unless condition
  # ...
else
  # the conditions that lead to this code
end
```

is super hard on your brain because it's logically 'not, not' the condition.

nodunayo: Okay, maybe you want to say something like:

You should definitely avoid using `unless` and `else` in conjunction in a conditional. This structure is particularly hard to reason about because the `else` block acts as a double negative in the conditional. Instead, replace the `unless` for a logically equivalent `if` and swap the code between the two branches of the conditional. (Probably a better way to say this, particularly the last bit..)

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -10,9 +10,9 @@ image:
   credit: 'Pablo Garcia Saldaña'
 ---
 
-Ruby has a bunch of delightful syntactic sugar. One example of this is `unless`, which you can use in place of an `if` with a negative.
+One of Ruby’s strengths is it’s delightful built-in syntactic sugar. One example of this is `unless`, which you can use in place of using `if` combined with a negative statement.
```
nodunayo: *its (not 'it's).

I love this new first sentence!

nodunayo: *is the _keyword_ `unless` (perhaps)

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -48,13 +48,15 @@ end
 
 ## But why?
 
-Ruby’s conditional syntax is ‘truthy’. In Ruby’s case that means `nil` is equivalent to `false`. There is no need to do a negated `#nil?` check on an object as it is redundant as shown in the first two examples.
+Ruby’s conditional syntax is ‘truthy’, meaning `nil` is equivalent to `false`. Therefore a negated `#nil?` check on an object is redundant.
 
-The syntax of `!!` in the third example converts a value to boolean, but it is unnecessary and likely obscures your intention. If you _really_ want to do a `nil` check, you should use `#nil?` instead.
+The syntax of `!!` is two consecutive boolean ‘not’ operators, if takes any value and converts it into `true` or `false`. However, given Ruby's ‘truthy’ conditionals it is unnecessary and adds redundant code.
```
nodunayo: I think there's a typo or two in this first sentence? Doesn't read properly.

nodunayo: And when you say "given Ruby's `truthy` conditionals, what you mean is that: in Ruby, values can be 'truthy' or 'falsey' in and of themselves, so you don't need to convert objects into a boolean?

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -48,13 +48,15 @@ end
...
 
-You should definitely avoid using an `else` block with an `unless`: the `else` block is a double negative of the condition. Instead, switch the `unless` for an logically equivalent `if` and swap the code in the successful conditional block for the code in the `else` block. This might require some refactoring to the condition to make the resulting code easy to understand.
+You should definitely avoid using `else` block with an `unless`. This structure is hard to reason about because the `else` block acts as a double negative of the condition. Instead, replace the `unless` for an logically equivalent `if` and swap the code in the two branches of the conditional.
```
nodunayo: I think you still want '_an_ `else` block'.

nodunayo: *_a_ logically equivalent

nodunayo: *swap the code _segments_ in the two branches of the conditional _around_

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -48,13 +48,15 @@ end
...
 
-It is also _much_ harder to reason when using an `unless` conditional with any boolean algebra (`&&` or `||`). It is an easy way to confuse yourself! The conditional `unless one && two` is the same code as `if !one || !two`. In this case the complex `if` condition is an opportunity to refactor the conditional into a well named method so it can be better understood.
+It is also _much_ harder to reason when using an `unless` conditional with any boolean algebra (`&&` or `||`). It is an easy way to confuse yourself! The conditional `unless one && two` is equivalent to `if !one || !two`. Although the logic looks messier I find the positive version easier to reason about.
```
nodunayo: I think change it from "I find..." to "people often find..."

FYI, this is the clearest of all of the "But why" paragraphs.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -48,13 +48,15 @@ end
 
 ## But why?
 
-Ruby’s conditional syntax is ‘truthy’. In Ruby’s case that means `nil` is equivalent to `false`. There is no need to do a negated `#nil?` check on an object as it is redundant as shown in the first two examples.
+Ruby’s conditional syntax is ‘truthy’, meaning `nil` is equivalent to `false`. Therefore a negated `#nil?` check on an object is redundant.
```
nodunayo: I think to make this whole section clearer you need to:

* explain upfront what you mean by "Ruby's conditional syntax is 'truthy'. Can you find another explanation that isn't just "meaning `nil` is equivalent to `false`." Because I feel like you need one more step in between before you get to the 'nil' bit. Does that make sense?

* tie each paragraph to one of the examples (like you had before?) but explain a bit more. For example, before you had something like: "it's redundant in the first two examples" but I think it'll be more powerful to confirm/explain what the non-redundant version looks like.

I know you're aiming for brevity here...but let's get all of the explanations clear then we can see where we can cut down.

Also, are there possibly two "One Ruby Things" in here? One around `nil` and `!!` and another around `unless`/`else` and using `if` with a positive?

_posts/2017/2017-12-31-avoid-double-negatives.md
nodunayo: I think you should just say that: `!!` conversts any value into a boolean of `true` or `false`. Don't mention the `not` operators thing. I think it confuses the sentence and takes away focus from your point/

nodunayo (dismissed):

_posts/2017/2017-12-31-avoid-double-negatives.md
andycroll: There were three.

nodunayo: Ahahaha.

_posts/2017/2017-12-31-avoid-double-negatives.md
andycroll: Or remove 'I find'.

nodunayo: Indeed!

Tell 'em how it is. :P

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+# Instead of Examples 1,2 & 3
+if something
+  # do something
+end
+```
+
+
+## But why?
+
+Ruby’s conditional syntax is ‘truthy’, meaning that anything that is `nil` is also considered to be `false` and anything not-`nil` can be considered to be `true`.
```
nodunayo: 👍

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+Performing a `#nil?` check in a negative conditional, as in the first two examples, is often redundant as any non-`nil` value is `true`. So you can reverse the conditional and remove the check for `nil` and end up with clearer code.
```
nodunayo: I wonder if it's clear that by 'reverse the conditional' you mean, either:
* if you have an `unless` change it to an `if`, or
* if you have a '!' operator, get rid of it..

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+The syntax of `!!`, in the third example, is a shorthand for turning any value (either ‘truthy’ or ‘falsey’) into the _actual boolean values_ `true` or `false`. However, given Ruby's ‘truthy’ conditionals you don’t need to perform this redundant conversion.
```
nodunayo: *is shorthand (remove 'a', I think...)

nodunayo: Maybe, something like: "However, given Ruby's conditionals evaluate each branch's statement as either 'truthy' or 'falsey', performing this conversion is redundant.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+Avoid using an `else` block with an `unless`. This structure is hard to reason about because the `else` block acts as a double negative of the condition. Instead, replace the `unless` for a logically equivalent `if` and swap the blocks of code in the two branches of the conditional around.
```
nodunayo: *the `else` block 'has to be interpreted' as a double negative.. or 'reads as a' double negative.

Not sure what to say here but "acts as a double negative of the condition" isn't clear.

nodunayo: I would get rid of 'logically equivalent' and explain what to do:
"Instead, replace the `unless` for `if` and swap around the blocks of code in the two branches of the conditional." (Note how I moved the word 'around' to earlier in the sentence...I think that's the correct way with respect to grammar, but not sure. Either way, I think it makes that sentence easier to read).

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+It is _much_ harder to reason when using an `unless` conditional with any boolean algebra (`&&` or `||`). It is an easy way to confuse yourself! The conditional `unless one && two` is equivalent to `if !one || !two`. Although the logic looks messier the positive version is easier to reason about.
```
nodunayo: *"...harder to work out what's going on" or "...harder to read/understand code"
I don't think "harder to reason" on its own is enough in this sentence.

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+A complex `if` condition might well be an opportunity to extract the conditional into a well-named method so it can be better understood.
```
nodunayo: *more easily understood. (A suggestion, only).

_posts/2017/2017-12-31-avoid-double-negatives.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Why not?
+
+I’m definitely not saying don't use `unless`, but it's important to remember it is _sugar_ and thus it is important only to use it where it makes the resulting code clearer.
```
nodunayo: *so it's important to use it only when it makes the resulting code clearer.

nodunayo (dismissed):

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,57 @@
...
+layout: article
+category: ruby
+image:
+  base: '2017/truthiness-in-conditionals'
+  alt: 'Truth'
+  source: 'https://unsplash.com/photos/Aa2h3S3E47k'
+  credit: 'Jarrod Fitzgearlds'
+---
+
+Ruby’s conditional syntax is ‘truthy’, meaning that anything that is `nil` is also considered to be `false` and anything not-`nil` can be considered to be `true`.
```
nodunayo: *meaning that any _statement in a conditional_ that evaluates to `nil`..?

nodunayo: *is also considered to be equivalent to `false`

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,57 @@
...
+# Instead of Examples 1,2 & 3
+if something
+  # do something
+end
+```
+
+
+## But why?
+
+Performing a `#nil?` check in a negative conditional, as in the first two examples, is often redundant as any non-`nil` value is `true`. You can remove the `nil?` check and substitute the `unless` for an `if` or remove the `!` and end up with clearer code that means the same thing.
```
nodunayo: *Performing a `#nil?` check in a negative conditional, as in the first two examples, is often redundant. This is because any non-`nil` value is truthy on its own. Instead, remove the `#nil?` check and substitute the `unless` for an `if` (for Example 1) or remove the `!` (for Example 2) and end up with clearer code that means the same thing.

nodunayo: Even with my suggested changes above...not completely happy with this paragraph...I keep wanting to add more words then I take them out because I think you lose the brevity.

I'll tell you the bits I'm not sure about:

* "...in a negative conditional" — is it clear what you mean here? You mean in a conditional branch that's checking a negative case? Or if you've got 'unless' or '!'-something? Are there any other 'negative' conditionals?

* "as any non-nil value is true" — I'm tempted to add another sentence to explain that the negative statement and the nil check cancel each other out...but then I think it's just getting long..and maybe people do already understand why it's redundant. I also don't think stating the equivalence of non-`nil` and `true` is that clear, hence my suggested edit..but maybe I'm over-thinking the words and how others will read and understand them.

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,57 @@
...
+
+The syntax of `!!`, in the third example, is shorthand for turning any value (either ‘truthy’ or ‘falsey’) into the _actual boolean values_ `true` or `false`. However, given Ruby's ‘truthy’ conditionals performing this conversion is redundant.
+
+
+## Why not?
+
+This comes down to understandability. If you _really_ are checking for `nil`, then by all means explicitly use the check.
```
nodunayo: Needs to be clearer what _really_ checking for `nil` means...I would read this and go: wait, when is okay to use the check?

_posts/2018/2018-01-07-be-sparing-when-using-unless.md
```diff
@@ -0,0 +1,66 @@
+---
+title: 'Be sparing when using unless'
+description: 'Gimme some sugar, but not too much'
```
nodunayo, line 3: Love it.

_posts/2018/2018-01-07-be-sparing-when-using-unless.md
```diff
@@ -0,0 +1,66 @@
...
+Avoid using an `else` block with an `unless`. This structure is hard to reason about because the `else` block is a double negative: it is "not not" the specified condition. Instead, replace the `unless` with an `if` while refactoring the condition and swap around the blocks of code in the main and `else` branches.
+
+It is _much_ harder to understand the code when using an `unless` conditional with boolean algebra (`&&` or `||`). It is an easy way to confuse yourself! The conditional `unless one && two` is equivalent to `if !one || !two`. Although the logic looks messier the positive version is easier to reason about.
+
+A complex `if` condition might well be an opportunity to extract the conditional into a well-named method so it can be more easily understood.
+
+
+## Why not?
+
+I’m definitely not saying don't use `unless`, but it's important to remember it is _sugar_ and thus its important only to use it where it makes the resulting code clearer.
```
nodunayo: *it's

nodunayo (dismissed):
Much better as two!!

I think the `unless` one reads slightly clearer..but the other one is nearly there...I'm not sure sure how much I'm over-complicating things though. See what you think.

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,59 @@
...
+# Instead of Examples 1,2 & 3
+if something
+  # do something
+end
+```
+
+
+## But why?
+
+Performing a `#nil?` check as part of a statement in a negative conditional, as in the first two examples (`unless` or `if !`), is often redundant. Any non-`nil` value is considered to be ‘truthy’.
```
nodunayo: *is often redundant as any  non-`nil` value is `truthy`. Therefore, you might as well perform the check as part of a statement in a positive conditional.

I still think there's something that's missing that is the: "oh, of course. That's why it's redundant".

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,59 @@
...
+
+Remove the `nil?` check and substitute the `unless` for an `if` (example 1) or remove the `!` (example 2) and end up with clearer code that means the same thing.
+
+The syntax of `!!`, in the third example, is shorthand for turning any value (either ‘truthy’ or ‘falsey’) into the _actual boolean values_ `true` or `false`. However, given Ruby's ‘truthy’ conditionals performing this conversion is redundant.
+
+
+## Why not?
+
+This comes down to understandability. If you _really_ are checking for `nil`, perhaps you're dealing differently with an empty array and `nil`, then by all means explicitly use the check.
```
nodunayo: * If you _really_ are checking for `nil` — perhaps you're want to deal with an empty array and `nil` in different ways — then, by all means, explicitly use the check.

_posts/2018/2018-01-07-be-sparing-when-using-unless.md
```diff
@@ -0,0 +1,66 @@
...
+# Example 2
+if !something? && !another_thing?
+  # do something
+end
+```
+
+
+## But why?
+
+Avoid using an `else` block with an `unless`. This structure is hard to reason about because the `else` block is a double negative: it is "not not" the specified condition. Instead, replace the `unless` with an `if` while refactoring the condition and swap around the blocks of code in the main and `else` branches.
```
nodunayo: * Instead, replace the `unless` with an `if`, taking care to refactor the condition statement as is necessary, and swap around...

nodunayo (dismissed):
Almost there...!

_posts/2017/2017-12-31-truthiness-in-conditionals.md
andycroll: Any `nil` value is 'false-y', so you can achieve the same result with a positive conditional and no need for the `#nil?` check.

nodunayo: Any nil value is 'falsey', so you can achieve the same result with a positive conditional and no #nil? check.

_posts/2017/2017-12-31-truthiness-in-conditionals.md
```diff
@@ -0,0 +1,59 @@
...
+Performing a `#nil?` check as part of a statement in a negative conditional, as in the first two examples (`unless` or `if !`), is often redundant. Any `nil` value is ‘falsey’, so you can achieve the same result with a positive conditional and no `#nil?` check.
+
+Remove the `nil?` check and substitute the `unless` for an `if` (example 1) or remove the `!` (example 2) and end up with clearer code that means the same thing.
+
+The syntax of `!!`, in the third example, is shorthand for turning any value (either ‘truthy’ or ‘falsey’) into the _actual boolean values_ `true` or `false`. However, given Ruby's ‘truthy’ conditionals performing this conversion is redundant.
+
+
+## Why not?
+
+This comes down to understandability. If you _really_ are checking for `nil` — perhaps you’re treating an empty array and `nil` in different ways — then by all means explicitly use the check.
```
nodunayo: *then, by all means, explicitly

_posts/2018/2018-01-07-be-sparing-when-using-unless.md
```diff
@@ -0,0 +1,66 @@
...
+
+However, given the flexibility of Ruby's syntax, it is easy to make code harder to understand than it needs to be.
+
+## Instead of…
+
+...using `unless` in ways that makes your brain hurt.
+
+```ruby
+# Example 1
+unless something?
```
nodunayo, line 23: I think you want to remove the `?` here, or add one on line 42, for continuity purposes?

andycroll: Added on 42.

nodunayo:
👌

PR #19, opened 2018-04-26
Multiple approaches to not doing work in jobs

andycroll:


_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/a-job-should-know-whether-to-run-itself'
+  alt: 'A queue'
+  source: 'https://unsplash.com/photos/Nzb4LBsctyQ'
+  credit: 'Hal Gatewood'
+---
+
+Like other `ActiveWhatever` objects in Rails, a job also comes with a selection of callbacks that are run at different points in it's lifecycle.
```
nodunayo: * Like other `Active`-prefixed objects in Rails, Active Job objects also come with a selection of...

Though, I'm also tempted to just have: Active Job objects come with a selection... as the starting.

Also "its", not "it's".

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+
+A list of the available callbacks are in the [Activejob Rails Guide](http://guides.rubyonrails.org/active_job_basics.html#callbacks).
```
nodunayo: *Active Job

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+
+We can use these callbacks to save a lot of conditional logic for jobs that might not always need to be run.
```
nodunayo: Are you saying we save ourselves time writing conditional logic for these jobs...or we can save writing conditional logic for the jobs that don't always need to be run...

I'm pretty sure you're saying the former, but for those unfamiliar with this the latter could be read.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+end
+
+# then
+DoItLaterJob.perform_later(thing)
+```
+
+
+# Or...
+
+...even throw the symbol `:abort` in a callback so a job does not even enqueue itself.
```
nodunayo: Take out one of the 'even's here...I think the first one should go.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+end
+
+# then
+DoItLaterJob.perform_later(thing)
+```
+
+
+## But why?
+
+It's generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing.
```
nodunayo: *It

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+In both cases the code where the enqueuing happens becomes clearer without the conditional. This is a case where either situation can be the correct path, it’s a matter of context.
+
+
+##  Why not?
+
+There may be a slight performance cost of enqueuing (and then dequeuing) jobs when using the 'early return' style. You might be filling your queue with lots of jobs that don’t do anything.
+
+There are also subtle logical & timing differences between the three approaches for 'not running a job'. It requires careful consideration of these factors to establish which choice to make.
+
+This is one of those cases where all three are valid approaches and you might even use them together for different conditions.
```
nodunayo: Does this fit in the 'Why not?' section?

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,92 @@
...
+
+## But why?
+
+It's generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing.
+
+Use the 'early return' style when the property might change before the job runs, For example, if you scheduled a job to run in the future and the property of the object the job affects might change.
+
+Use the 'callback and throw' style when the enqueuing should be based on a condition that is unlikely to change.
+
+In both cases the code where the enqueuing happens becomes clearer without the conditional. This is a case where either situation can be the correct path, it’s a matter of context.
```
nodunayo: But all three have a conditional...why are the other two clearer?

Also, hypen or colon after 'correct path'.

nodunayo (dismissed):

nodunayo:
I think this post is missing a concrete example or some explanation of when you'd want to use Active Job and what sort of thing 'thing' is...

If you could find one concrete example to run through this post I think it would take it to the next level.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
andycroll: Both.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/a-job-should-know-whether-to-run-itself'
+  alt: 'A queue'
+  source: 'https://unsplash.com/photos/Nzb4LBsctyQ'
+  credit: 'Hal Gatewood'
+---
+
+It is a good idea to do as much of the ‘work’ of an application in asynchronous jobs as it means doing less work in each controller action. Rails includes `ActiveJob` as a way to implement this pattern in your application.
```
nodunayo: *as much of the 'work' of an application _as possible_

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+
+Each `ActiveJob` comes with a selection of callbacks that are run at different points in its lifecycle.
```
nodunayo: *in a job's lifecycle

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+
+A list of the available callbacks are in the [ActiveJob Rails Guide](http://guides.rubyonrails.org/active_job_basics.html#callbacks).
```
nodunayo: *Active Job

If you check the docs, etc., outside of the codebase it's referred to as two separate words.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+
+When a job doesn't always need to be run, we can use these callbacks to save writing conditional logic every time we enqueue one.
+
+
+## Instead of…
```
nodunayo: *as you enqueue it (?)

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+# then
+SendNotificationJob.perform_later(user, message)
+```
+
+
+## But why?
+
+It generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing.
+
+Use the 'early return' style when the property might change before the job runs, For example, if you scheduled a job to run in the future and the property of the object the job affects might change.
```
nodunayo: This paragraph is confusing.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+
+Use the 'callback and throw' style when the enqueuing should be based on a condition that is unlikely to change.
+
+In both cases the code the code for the enqueuing is clearer without the conditional. The enqueuing code doesn’t need to concern itself with whether the work the job needs to be done or not, the job itself can decide.
```
nodunayo: 'the code' is repeated twice.

nodunayo: *without the _external_ conditional?

Because the other two examples still have conditionals...

Also — what happens in the first case when the job is already enqueued? Does it just get later removed if `perform_later` is never called? (Just wondering for me).

nodunayo: The second sentence of this paragraph is also not very clear.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+##  Why not?
+
+There may be a slight performance cost of enqueuing (and then dequeuing) jobs when using the 'early return' style. You might be filling your queue with lots of jobs that don’t do anything.
+
+There are also subtle logical & timing differences between the three approaches for 'not running a job'. It requires careful consideration of these factors to establish which choice to make.
```
nodunayo: This `Why not?` section leaves me wondering if I should even bother changing how I'm doing things if I'm already doing the `Instead of` version.

Normally in your posts I come to the end thinking: okay, I should definitely look into this thing/consider changing how I do this thing, etc..but with this one, I'm thinking: sooo, I don't really need to change anything..? Or: I'm not clear when I would want to change how I use Active Job.

nodunayo (dismissed):

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
andycroll: > It generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing.

This is kinda the thing. You want to keep the logic of 'when to run' near the 'what to do'.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
andycroll: 🤓

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+image:
+  base: '2018/a-job-should-know-whether-to-run-itself'
+  alt: 'A queue'
+  source: 'https://unsplash.com/photos/Nzb4LBsctyQ'
+  credit: 'Hal Gatewood'
+---
+
+It is a good idea to do as much of the ‘work’ of an application as possible in asynchronous jobs as it means doing less work in each controller action. Rails includes `ActiveJob` as a way to implement this pattern in your application.
+
+Each `ActiveJob` comes with a selection of callbacks that are run at different points in in a job’s lifecycle.
```
nodunayo: Get rid of one of the 'in's. :-)

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+
+A list of the available callbacks are in the [Active Job Rails Guide](http://guides.rubyonrails.org/active_job_basics.html#callbacks).
+
+When a job doesn't always need to be run, we can use these callbacks to save writing conditional logic in every location in the code we enqueue the job.
```
nodunayo: *_where_ we enqueue the job (?)

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+# then
+SendNotificationJob.perform_later(user, message)
+```
+
+
+## But why?
+
+It generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing. The code enqueuing the job doesn’t need to care whether the work of the job needs to be done or not.
+
+The job can encapsulate the 'whether it should happen' alongside the 'how/what', this will make the code easier to understand when you come back to it later.
```
nodunayo: * Full stop after 'how/what'.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+end
+
+# then
+SendNotificationJob.perform_later(user, message)
+```
+
+
+## But why?
+
+It generally aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing. The code enqueuing the job doesn’t need to care whether the work of the job needs to be done or not.
```
nodunayo: I think rework this paragraph and the next one as follows:

The code enqueuing the job doesn’t need to care whether the work of the job needs to be done or not. It aids comprehension to keep the logic about _when_ to do something near to the code that actually does the thing. This will make the code easier to understand when you come back to it later.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+
+The job can encapsulate the 'whether it should happen' alongside the 'how/what', this will make the code easier to understand when you come back to it later.
+
+A job happens an unspecified period of time after it is enqueued. It's possible the answer to 'should this job be run?' could change between the enqueuing and the execution. This is a good candidate for the first, 'early return', style.
+
+The second, 'use throw in a callback', style is logically closer to the initial condition as the job is never enqueued.
```
nodunayo: I get what you mean, but maybe there's some rewording here to explain what you mean?

Now that you've reworked when to use the 'early return' version, I think saying something like the' use throw in a callback' style is for when the condition for running a job is very unlikely to change between enqueuing and execution.

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,98 @@
...
+
+##  Why not?
+
+The code for the ‘use throw in a callback’ style is slightly more complex than an external conditional when you enqueue.
+
+Also it's only worth encapsulating the 'should the job be run' logic if the condition is always applied to the job.
+
+There may be a slight performance cost of enqueuing (and then dequeuing) jobs when using the 'early return' style. You might be filling your queue with lots of jobs that don’t do anything.
+
+This is one of those cases where both ‘not enqueuing at all’ and ‘returning early’ are valid approaches and you _might_ even use them together for different conditions.
```
nodunayo: Why is this last paragraph in the `Why not?` section?

nodunayo (dismissed):

_posts/2018/2018-05-13-a-job-should-know-whether-to-run-itself.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+
+## But why?
+
+The code enqueuing the job doesn’t need to care whether the work of the job needs to be done or not. It aids comprehension to keep the logic about when to do something near to the code that actually does the thing. This will make the code easier to understand when you come back to it later.
+
+A job happens an unspecified period of time after it is enqueued. It's possible the answer to 'should this job be run?' could change between the enqueuing and the execution. This is a good candidate for the first, 'early return', style.
+
+The second, 'use throw in a callback', style is logically closer to the initial condition as the job is never enqueued. It is sensible to use this style when the condition for running a job is very unlikely to change between enqueuing and execution.
```
nodunayo: *It is sensible to use the second, 'use throw in a callback', style when the condition for running a job is very unlikely to change...execution. This is because, with this style, the job is never enqueued if the initial condition isn't met.

nodunayo (approved):
I still don't like the the first sentence on line 83....I just don't think it's as clear as the rest of the piece. But feel free to ignore my suggestion for a re-write. 😁

PR #20, opened 2018-05-17
First and last don't mean what you think

andycroll:
Bit late. But blessedly short.


_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/first-and-last-may-not-mean-what-you-think'
+  alt: 'Running track'
+  source: 'https://unsplash.com/photos/52p1K0d0euM'
+  credit: 'Austris Augusts'
+---
+
+Each ActiveRecord model comes with the well used ‘get me a single record’ scopes; `.first` & `.last`.
```
nodunayo: * Active Record

nodunayo: *well-used

nodunayo: Colon instead of semi-colon?

nodunayo: Also, use 'and' instead of an ampersand, methinks..?

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+  # ...
+end
+
+User.most_recently_created
+```
+
+
+## But why?
+
+I’ll admit, this is a subtle one. My argument is mostly about making the code more explicit.
```
nodunayo: I think there's a better (subtly different) way to say this sentence, but maybe not: My argument is mostly about making the code more explicit.

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+
+If you’re using the bare `.first` and `.last` methods on an bare ActiveRecord model scope you are relying heavily on the Rails convention of an incrementing integer ID.
```
nodunayo: Take out the first 'bare'.

nodunayo: *on _a_ bare

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+
+Those methods are thus colloquially used to mean ‘created first’ and ‘created most recently’. It's only an accident of Rails default design that sorting by ascending `id` is same order as the ascending `created_at` timestamp.
```
nodunayo: I get what you are saying here but I think you can make this slightly more straightforward.

I think there's some remixing involved between this paragraph and the one above. You want to first talk about what people think the '.first' and '.last' methods mean — and then go into how that falsely relies on the incrementing integer and the fact that it's the same ascending order. Does that make sense?

It seems as if there's a story in these two paragraphs/a thought process, that's slightly jumbled up.

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+
+You might remember I advocated [using UUIDs for primary keys](/2017/choose-uuids-for-model-ids-in-rails) in a previous article, this pairs up nicely don’t you think?
```
nodunayo: * in a previous article. (full stop) This pairs up...

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+### Why not?
+
+It’s more code and Rail’s conventions on this are long standing. You might not feel this is worth it.
```
nodunayo: *Implementing this leads to more code and _Rails'_ conventions on this are long-standing.

nodunayo (dismissed):

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+
+## But why?
+
+This is mostly about making the code more explicit.
+
+The `.first` and `.last` methods, when called on a bare Active Record model scope, are colloquially used to mean ‘created first’ and ‘created most recently’.
+
+It’s only an accident of Rails default design that sorting by ascending `id` is the same order as the ascending `created_at` timestamp.
```
nodunayo: *Rails'

nodunayo: What you're actually trying to say here is that:  first returns the smallest number ID, and last returns the largest number ID. As you create objects, Rails assigns IDs incrementally. Therefore, it's only by luck that when you call 'first' you get the 'earliest created' — really you're asking Rails for: give me the lowest number ID.

If that's correct, I think that there must be a way to take what I've said above and improve your paragraph — not quite sure what that improvement is yet.

Maybe even merge in the paragraph before where you refer to the 'colloquial' use of 'first' and 'last'.

_posts/2018/2018-05-27-first-and-last-may-not-mean-what-you-think.md
```diff
@@ -0,0 +1,58 @@
...
+
+Using these scopes with no explicit order means you are relying heavily on the Rails convention of an incrementing integer `id`.
+
+You might remember I advocated [using UUIDs for primary keys](/2017/choose-uuids-for-model-ids-in-rails) in a previous article. This pairs up nicely don’t you think?
```
nodunayo: With your "This pairs up nicely don't you think?" are you saying: if using UUIDs, this throws this whole 'first' and 'last' thing to shit?

Maybe get rid of that last question, and depending on how the preceding paragraph ends change it to something like: For example, imagine if you took my advice from a previous article where I advocated using UUIDs for primary keys?

Just a suggestion — tricky to see how it'll fit together without previously suggested edits.

nodunayo (dismissed):
More subtle/not fully-formed suggestions. See what you think!

nodunayo:
Yeah — so I just noticed the title of this post: First and last don't mean what you think.

I realise one of the things that doesn't feel right about this post — I don't think you explicitly say what 'first' and 'last' do mean — which I think I allude at in one of my comments. I think you should do this for clarity.

PR #25, opened 2018-06-11
`\z` should be preferred over `\Z` to match end of string

cbliard:
Hi Andy,

There is a mix of `\z` and `\Z` in last post. `\z` should be preferred as it matches only the end of the string, whereas `\Z` can also match if there is a final newline after it.

This pull request uses `\z` everywhere.

```ruby
> "word".match?(/\A[a-z]+\z/)
=> true
> "word".match?(/\A[a-z]+\Z/)
=> true
> "word\n".match?(/\A[a-z]+\z/)
=> false
> "word\n".match?(/\A[a-z]+\Z/)
=> true
> "word\n\n".match?(/\A[a-z]+\z/)
=> false
> "word\n\n".match?(/\A[a-z]+\Z/)
=> false
```

### DIFF
--- a/_posts/2018/2018-06-10-beginning-and-end-of-string-in-regex.md
+++ b/_posts/2018/2018-06-10-beginning-and-end-of-string-in-regex.md
@@ -30,12 +30,12 @@ However, we should beware.
 
 ## Use…
 
-…`\A` and `\Z`.
+…`\A` and `\z`.
 
 ```ruby
 # A regular expression matching a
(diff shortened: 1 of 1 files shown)


andycroll (approved):
You're absolutely right.

And actually that's what I meant to type! I'm not entirely sure how I missed that typo.

You're a star. Thank you!

cbliard:
My pleasure :)

PR #27, opened 2018-08-12
Presence method, first draft

andycroll:


_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/use-the-presence-method'
+  alt: 'Ghostly'
+  source: 'https://unsplash.com/photos/ZC0EbdLC8G0'
+  credit: 'Stefano Pollio'
+---
+
+Active Support takes a lot flak for the magic it adds to core Ruby libraries including patching a lot of functionality into Ruby's `Object`.
```
nodunayo: Maybe reword this to say: Active Support takes a lot of flat because it adds a lot of functionality to core Ruby libraries, including patching a lot of functionality into Ruby's `Object` _class_.

Also, is there a way you can avoid using 'functionality' twice so close together?

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+The documentation can be found [here](https://api.rubyonrails.org/classes/Object.html#method-i-presence).
```
nodunayo: An ex-colleague once put me off the `here` hyperlink method for sharing links. Their argument was that it wasn't semantic or accessible. So..if you wanted to subscribe to that thinking, you can add a line to the end of the preceding paragraph saying: "Have a look at [the documentation]() to find out more." Or something like that...

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+For me, a lot of the delight in Rails comes from the shortcuts it adds. Some features even eventually make their way into Ruby core.
```
nodunayo: This last sentence feels misplaced. I would rather add another sentence (or extend the first one) to indicate that you therefore are a fan of Active Support (and maybe add why).

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+class User < ApplicationRecord
+  validates :email, presence: true
+
+  def full_name
+    if nickname.present?
+      nickname
+    elsif given_name.present?
+      given_name
+    else
+      email.split('@').first
```
nodunayo, line 36: This method seems to be sourcing `first_name` rather than the `full_name`?

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+end
+```
+
+## But why?
+
+This is a method I don't see used as much as the more familiar `#blank?` and `#present?`.
+
+It provides a very convenient shortcut to return either the object (if it exists) or `nil`.
+
+It is the equivalent of writing `object.present? ? object : nil` which is pattern you often see in Ruby code, where `nil` must often be dealt with.
```
nodunayo: - Add a comma before 'which is a pattern'
- Add a full stop after 'Ruby code'
- Then edit the last sentence to explain how you then have to deal with potential `nil` values

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+### Why not?
+
+This `#presence` method is only available in Rails, if you're just using standard Ruby you might not want to include Active Support just for this.
```
nodunayo: Add a full stop after 'Rails'.

Then: If you're _only_ using Ruby, it's probably not worth including Active Support just for this. (Or something like that)

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+Even when using Rails there can be understandable resistance to Rails' habit of extending standard Ruby classes.
+
+Altering standard classes in your own code, by 'monkey patching', can cause bugs when the standard Ruby classes begin to behave in unexpected ways.
```
nodunayo: Is this paragraph to say: even in your own code people don't like monkey patching so it's understandable you wouldn't like it in the framework you're using?

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+Your level of aversion to side-effects might be enough to avoid this particular piece of Rails' magic.
```
nodunayo: But what are the potential side-effects? You haven't really expressed what could go wrong if I start using this extra functionality in Active Support.

Also — why doesn't the intro refer to the `presence` method? After all — that is what this article is about...

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,80 @@
...
+
+That said... I consider all of Rails, as a diligently maintained and widely used set of libraries, to be fair game for use if it makes the resulting code clearer.
```
nodunayo: No need for the ellipses.

Maybe edit sentence to read: "Given Rails is a diligently maintained and widely used set of libraries, I consider all and any of it to be fair game to use if it makes the resulting code clearer."

Maybe you want to change 'fair game' to something like 'safe' or 'shrewd', etc..something that gets at either people's concern about side effects or the fact that you think: it's all there and well-tested, would be a shame/waste not to use it, since you're already on the Rails train!

nodunayo (dismissed):

_posts/2018/2018-08-19-use-the-presence-method.md
andycroll: "Rails train" is a double metaphor :clap:

_posts/2018/2018-08-19-use-the-presence-method.md
andycroll: Not exactly.

```ruby
'nodunayo@email.com'.split('@').first
# -> 'nodunayo'
```

plus it's a fallback :-)

nodunayo: But ‘nickname’ and ‘given name’ are all examples of the first name, right? I’m saying, even with your fallback, a more suitable name for this method seems to be ‘first_name’, rather than ‘full_name’.

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,82 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/use-the-presence-method'
+  alt: 'Ghostly'
+  source: 'https://unsplash.com/photos/ZC0EbdLC8G0'
+  credit: 'Stefano Pollio'
+---
+
+Active Support takes a lot flak for because it adds methods to core Ruby libraries, most notably patching a lot of functionality into Ruby's `Object` class.
```
nodunayo: Stray 'for' in there

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,82 @@
...
+
+Every object in Ruby is a subclass of `Object`, thus if you add methods to it, you're adding them to _every_ object in your code.
+
+Have a look at the documentation [for the enhancements in Active Support](https://api.rubyonrails.org/classes/Object.html#method-i-presence) to find out more.
```
nodunayo: I would extend the hyperlink to cover 'the documentation'.

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,82 @@
...
+end
+```
+
+## But why?
+
+Use of the `#presence` method provides a very convenient shortcut to return either the object (if it exists) or `nil`.
+
+It is the equivalent of writing `object.present? ? object : nil`, which is a pattern you often see in views in Rails where data may or may not exist.
+
+It's also a nice solution when you have an empty string or array, in those cases `#presence` returns `nil`.
```
nodunayo: Full stop after 'array'.

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Why not?
+
+This `#presence` method is only available in Rails. If you're only using Ruby it's probably not worth including Active Support just for this.
+
+Even when using Rails there can be understandable resistance to Rails' habit of extending standard Ruby classes.
+
+Altering standard classes in your own code, by [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch), often causes bugs when the standard Ruby classes begin to behave in unexpected ways. This is the main source of the concern about this style of coding.
+
+Your level of concern about the pitfalls of monkey patching _might_ be strong enough to avoid it even when Rails does it.
```
nodunayo, line 80: Great way of summarising this pain point.

_posts/2018/2018-08-19-use-the-presence-method.md
```diff
@@ -0,0 +1,82 @@
...
+
+Given Rails is a diligently maintained and widely used set of libraries, I consider all and any of it to be safe to use if it makes the resulting code clearer.
```
nodunayo, line 82: Great! :-)

PR #29, opened 2018-08-26
Using any? over present?

andycroll:


_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/careful-with-present-when-using-activerecord-scopes'
+  alt: 'Stadtbibliothek, Stuttgart, Deutschland'
+  source: 'https://unsplash.com/photos/PkbZahEG2Ng'
+  credit: 'Tobias Fischer'
+---
+
+We’re trained when using Rails to use `#present?` all over the place. For example, you might use it before trying to loop through an array in a view.
```
nodunayo: *When using Rails, we're trained to use `#present?` all over the place.

Side question: Are we? Is 'trained' the right word? It might be...I'm just wondering myself about its prevalence and how people perceive the method.

nodunayo: On the second sentence, that starts with 'For example', you need to explain a little more about why exactly you'd be using `#present?` in that case.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+However using `#present?` on models with large datasets can create unforeseen problems.
+
+
+## Instead of…
+
+...using `#present?` on an Active Relation (a uninstantiated scope on an Active Record class)
```
nodunayo: "Active Relation (a uninstantiated scope on an Active Record class)" — it wasn't immediately clear what you meant. Can you say this differently?

One of the main issues with it is the use of the word 'uninstantiated'. Do you mean undefined, or there's nothing in the scope? I know you don't mean this — but I think some people might pause and try and figure out what you're talking about.

* It'd also have to be _an_ uninstantiated... if that word stays.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+books = Book.recently_released
+if books.any?
+  books.paperbacks.each { |book| puts book.title }
+end
+```
+
+
+## But why?
+
+When you call `#present?` it instantiates the models in an array in memory before checking whether that array is empty or not. In most cases this will be fine but in extreme cases, say complex models, or a large dataset you might see large memory usage and slowdowns.
```
nodunayo: *the objects (rather than 'the models')?

*..will be fine, but in extreme cases, say complex models or when using a large dataset, you might see cumbersome (or another word) memory usage and performance impacts (perhaps, instead of 'slowdowns') (or 'a slower performing app' or something like that.)

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+If you use `#any?` a different, simpler SQL command is run that doesn't build a large array of Active Record objects.
```
nodunayo: *If you use `#any?`, (comma) a different, simpler....

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+There’s also no penalty to using it if the records are already loaded.
```
nodunayo: In what circumstances would that be? And surely I can then just use `#present?`. I think this line needs more elaboration.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+As always with changes to how your application uses the database, you’ll definitely want to monitor how the changes affect the real-life performance characteristics of _your_ application.
```
nodunayo: Are you implying that for some people the change won't make a difference/be worse? Maybe spell that out and the cases where that might prove to be true.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+You _could_ even monkey-patch the default behaviour of `#present?` on Active Relations in Rails by using the [`attendance` gem](https://github.com/schneems/attendance). This is a bit of a blunt instrument though and you’ll want to watch the performance of your app very closely if you take the plunge.
```
nodunayo: Why would I choose use this gem? You haven't explain why I would.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+### Why not?
+
+Using `#any?` _is_ an extra SQL query, and the performance of your app might not benefit from the change in approach.
```
nodunayo: Extra than if you used `#present?`.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+Additionally the preloading of records might actually be more efficient if you then go on to use them.
```
nodunayo: * Remove 'actually'

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+A similar “fix” to that in the `attendance` gem, changing how `#present?` works on an Active Relation, has been [merged](https://github.com/rails/rails/pull/10539) and then [reverted](https://github.com/rails/rails/commit/2b763131eacaae5bff9ffb5015fbf367d594dc64) from Rails in favour of the existing behaviour. So there are definite performance subtleties to checking the presence of records.
```
nodunayo: The language here needs tightening up a little.

If you explain the `attendance` gem earlier more clearly, then I don't think you need to repeat here what it does. Then the last sentence might be better worded: "This perhaps implies/suggests that there are definite performance subtleties...What the Rails team thought might have been an obvious performance improvement may not have been in all cases." Or something like that. Maybe you don't need the last sentence I added.

nodunayo (dismissed):

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/careful-with-present-when-using-activerecord-scopes'
+  alt: 'Stadtbibliothek, Stuttgart, Deutschland'
+  source: 'https://unsplash.com/photos/PkbZahEG2Ng'
+  credit: 'Tobias Fischer'
+---
+
+When using Rails, we tend to use `#present?` in many places as it is added to _all_ Ruby objects by Rails. For example, you might use it in a view to check if an array has some elements before trying to loop through it and display some data.
```
nodunayo: I would say: as it is available on all Ruby objects when you're using Rails (or something like that)

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+However using `#present?` on models with large datasets can create unforeseen problems.
```
nodunayo: Comma after 'However'

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+books = Book.recently_released
+if books.any?
+  books.paperbacks.each { |book| puts book.title }
+end
+```
+
+
+## But why?
+
+When you call `#present?` it instantiates the objects into an array in memory before checking whether that array is empty or not. In most cases this will be fine but in extreme cases, say complex models or when using a large dataset, you might see very high memory usage and slow performing requests.
```
nodunayo: This needs a bit more clarification, since, if the 'array' is empty then no objects will be instantiated, right?

So in the example you give, if there are no `recently_released` books, what is instantiated when you call `#present?` on `books` — is it all the objects in `Book`?

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+If you use `#any?`, a different, simpler SQL command is run that doesn't build a large array of Active Record objects in memory.
+
+If the relation has already been used, meaning the objects are loaded into memory, there’s no penalty to using `#present?` as it uses the existing array rather than loading a new one.
+
+As always with changes to how your application uses the database, you’ll definitely want to monitor how the changes affect the real-life performance characteristics of _your_ application. You need to make sure the extra SQL request from the call to `#any?` isn't a worse performing solution than using `#present?` for your data.
+
+You _could_ even monkey-patch the default behaviour of `#present?` on Active Relations in Rails, to always use `#any?`, by using the [`attendance` gem](https://github.com/schneems/attendance). This is a bit of a blunt instrument though and you’ll want to watch the performance of your app very closely if you take the plunge.
```
nodunayo: Is this for people who prefer writing `#present?` or think it's semantically better? Why use `attendance` instead of just using `#any?`?

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+### Why not?
+
+Using `#any?` _is_ an extra SQL query, over using `#present?`, and the performance of your app might not benefit from the change in approach. The preloading of records might be more efficient if you then go on to use them.
```
nodunayo: *When you use `#any?` you are making an extra SQL query and the performance of your app...

(Get rid of the 'over using '#present?'' bit)

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+A similar “fix” to that in the `attendance` gem, changing how `#present?` works on an Active Relation, has been [merged](https://github.com/rails/rails/pull/10539) and then [reverted](https://github.com/rails/rails/commit/2b763131eacaae5bff9ffb5015fbf367d594dc64) from Rails in favour of the existing behaviour, where you have to be more specific and choose to use `#any?`.
```
nodunayo: * A similar "fix" to that offered/suggested/shown (etc.) in the `attendance` gem _was_ merged and  then reverted...

nodunayo (dismissed):

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
andycroll: It's more that if there are _lots_ and then you only use some (or don't use them at all).

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
andycroll: Saves hunting down every occurrence.

nodunayo: I thought of that...but then I thought 'Find & Replace' — but maybe that's dangerous...as you might have some other occurrence of 'present?' that's not a method call? 🤔

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/careful-with-present-when-using-activerecord-scopes'
+  alt: 'Stadtbibliothek, Stuttgart, Deutschland'
+  source: 'https://unsplash.com/photos/PkbZahEG2Ng'
+  credit: 'Tobias Fischer'
+---
+
+When using Rails, we tend to use `#present?` in many places as it is available on _all_ Ruby objects when you’re using Rails. For example, you might use it in a view to check if an array has some elements before trying to loop through it and display some data.
```
nodunayo: Oops, forgot that we start this sentence with "When using Rails".

Therefore, let's finish it with "...within the framework" or something like that.

Alternatively: When using Rails, '#present?' is available on _all_ Ruby objects. Therefore, we tend to use it in many places.

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+books = Book.recently_released
+if books.any?
+  books.paperbacks.each { |book| puts book.title }
+end
+```
+
+
+## But why?
+
+When you call `#present?` it instantiates the objects into an array in memory before checking whether that array is empty or not. In most cases this will be fine but in extreme cases, say complex models or when using a large dataset, you might see very high memory usage and slow performing requests. This happens whether you then use those objects or not.
```
nodunayo: When you call `#present?` on an Active Relation it loads all of the objects in that Active Record class  into an array in memory before checking whether the scoped/Active Relation array is empty or not.

What do you think?

_posts/2018/2018-09-02-careful-with-present-when-using-activerecord-scopes.md
```diff
@@ -0,0 +1,61 @@
...
+
+You _could_ even monkey-patch the default behaviour of `#present?` on Active Relations in Rails, to always use `#any?`, by using the [`attendance` gem](https://github.com/schneems/attendance) to save yourself hunting down every use of `#present?`. This is a bit of a blunt instrument though and you’ll want to watch the performance of your app very closely if you take this blanket approach.
+
+
+### Why not?
+
+When you use `#any?` you are making an extra SQL query and the performance of your app might not benefit from the change in approach. The preloading of records might be more efficient if you then go on to use them.
+
+A similar “fix” to that offered in the `attendance` gem, changing how `#present?` works on an Active Relation, was [merged](https://github.com/rails/rails/pull/10539) and then [reverted](https://github.com/rails/rails/commit/2b763131eacaae5bff9ffb5015fbf367d594dc64) from Rails in favour of the existing behaviour, where you have to be more specific and choose to use `#any?`.
```
nodunayo: I would change the commas around "changing how 'present?' works on an Active Relation" to hyphens or put it in parentheses.

nodunayo (approved):

PR #32, opened 2018-10-11
A scope returns a scope

andycroll:


andycroll:
Very short - still might miss Monday ('cos I'm so late). No big deal - no rush.

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/a-scope-should-return-a-scope'
+  alt: 'Coin-operated (tele)scope'
+  source: 'https://unsplash.com/photos/K40awqK-hTI'
+  credit: 'Laurenz Kleinheider'
+---
+
+The more you can stay on the ‘rails’ when coding Ruby on Rails application the easier you life will be when maintaining the apps you’re building.
```
nodunayo: *when coding a Ruby on Rails application 

OR

*when coding Ruby on Rails applications

nodunayo: *_your_* life

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+
+A good way of doing this is to try and stick to the patterns set out in the standard Rails APIs.
```
nodunayo: Do you want to include anything about scopes in your intro? I think the intros are most effective when you introduce the topic a little and make people wonder: oh yeah, what is the right thing to do in this case? Or when you introduce the specific pain point.

What do you think?

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+  scope :sent, -> { where.not(sent_at: nil) }
+  scope :recently_sent, -> { sent.order(sent_at: :desc) }
+  scope :most_recently_sent, -> { recently_sent.first } # returns an object or nil
+end
+```
+
+
+## Always…
+
+...return an `ActiveRelation` from a scope.
```
nodunayo: And maybe add "and use that in a method" — or something like that.

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+  def most_recently_sent
+    recently_sent.first
+  end
+end
+```
+
+
+## But why?
+
+This is more a case of code organisation rather than any change in how you’ll actual use your models.
```
nodunayo: *This change improves the organisation of your code, rather than changing how you'll use your models.

*actually (but we're maybe getting rid of it anyway soo... :-D)

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+
+When you call `.where` or `.order` it returns an Active Relation ‘scope’ that can be chained together with further scopes. You encourage flexibility and reuse if you maintain this behaviour where your own named scopes are able to be chained with others.
```
nodunayo: *where _all_ of you own named scopes...

_posts/2018/2018-10-14-a-scope-should-return-a-scope.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+### Why not?
+
+This is just a helpful heuristic to organise your code. If you don't feel the organisational benefit, simply don’t use it.
```
nodunayo: *If the organisational benefit doesn't seem useful to you...

Or maybe:

* If it doesn't seem like there's any organisational benefit to doing this...

Because your current version implies there is organisational benefit and you either see it or you don't...but if there is one, it should be a no-brainer to adopt the heuristic.

Also, I can't help but feel that there must be some other benefit to having all of your named scope behave in a consistent way...beyond just — your code is organised better — no?

nodunayo (dismissed):

PR #34, opened 2018-11-28
TRUE or FALSE

andycroll:


_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/always-force-booleans-to-be-true-or-false'
+  alt: 'Lovehearts, be true'
+  source: 'https://unsplash.com/photos/ohqX8M_z85E'
+  credit: 'Obi Onyeador'
+---
+
+Migrations, in common with much of Rails, are ‘[sharp knives](https://rubyonrails.org/doctrine/#provide-sharp-knives)’. Easy to use but potentially dangerous and you can cut yourself.
```
nodunayo:
```suggestion
Migrations, in common with much of Rails, are ‘[sharp knives](https://rubyonrails.org/doctrine/#provide-sharp-knives)’ — easy to use but potentially dangerous and you can cut yourself.
```

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+
+One way a migration can ‘cut’ you is by letting you make quick decisions about your database structure without considering the consequences!
+
+
+## Instead of…
```
nodunayo, line 18: Yay, you stopped the full stops! 😁😇

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+  def change
+    add_column :characters, :satisfied, :boolean
+  end
+end
+```
+
+
+## Always…
+
+...use constraints that prevent values from being `nil` that should be only `true` or `false`.
```
nodunayo: * you refer to `NULL` and `nil` across the two statements — might be easier to refer to one or explain the distinction.

nodunayo:
```suggestion
...use constraints that prevent values that should be only ever be `true` or `false` from ending up as `nil` .
```

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+  end
+end
+```
+
+
+## But why?
+
+This is a known issue called the tri-state problem. Ideally you don't want a boolean (which by definition should be `true` or `false`) to ever equal `nil`.
+
+Although `nil` behaves a lot like `false` in Ruby it isn't quite the same. Having a third state that you can think of as “a bit like `false`” really complicates the logic of your application.
```
nodunayo:
```suggestion
Although `nil` behaves a lot like `false`, in Ruby they are not quite the same. Having a third state that you can think of as “a bit like `false`” really complicates the logic of your application.
```

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+```ruby
+Character.where(satisfied: false)
+```
+
+
+### Why not?
+
+If you’re starting from scratch, just do this.
+
+If you already have this issue, and heaven knows I’ve done it, you will have to put together a careful migration strategy. This might also be true of adding these sorts of columns to existing tables, depending on your underlying database.
```
nodunayo:
```suggestion
If you already have this issue — and heaven knows I’ve fallen into the trap — you will have to put together a careful migration strategy. This might also be true of adding these sorts of columns to existing tables, depending on your underlying database.
```

nodunayo: The "this might also be true..." part of this sentence isn't clear. What might also be true? That you'll need a careful migration strategy when adding a column to an existing table compared to when you're changing a column on an existing table. 

Maybe you can say: "If you already have this issue...put together careful migration strategies for when you're changing columns or adding columns to existing tables.." Something like that.

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+
+Setting defaults and constraints on existing tables to avoid downtime or long running migrations needs careful thought.
```
nodunayo: This sentence appears randomly tacked on at the end. It's not clear why it's relevant to the rest of the post.

nodunayo (dismissed):

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
andycroll: Funny - I had it round this way before making the change to this ordering. Was right the first time!

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+layout: article
+category: ruby
+image:
+  base: '2018/always-force-booleans-to-be-true-or-false'
+  alt: 'Lovehearts, be true'
+  source: 'https://unsplash.com/photos/ohqX8M_z85E'
+  credit: 'Obi Onyeador'
+---
+
+Migrations, in common with much of Rails, are ‘[sharp knives](https://rubyonrails.org/doctrine/#provide-sharp-knives)’ — easy to use but potentially dangerous and you can cut yourself.
```
nodunayo:
```suggestion
Migrations, in common with much of Rails, are ‘[sharp knives](https://rubyonrails.org/doctrine/#provide-sharp-knives)’ — easy to use, but potentially dangerous, and you can end up cutting yourself.
```

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+
+One way a migration can ‘cut’ you is by letting you make quick decisions about your database structure without considering the consequences. In this case we'll examine the case of `NULL` in a boolen field your SQL (converted to `nil` when you use it in Ruby).
```
nodunayo: *boolean field _of_ your SQL _database_(?)

nodunayo: I would take the last bit out of parentheses and just say: "..SQL. This is converted to `nil` when you use it in Ruby."

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+  def change
+    add_column :characters, :satisfied, :boolean, default: false, null: false
+  end
+end
+```
+
+
+## But why?
+
+This is a known issue called the tri-state problem. Ideally you don't want a boolean (which by definition should be `true` or `false`) to ever be `NULL` (or `nil` when it is converted from SQL to Ruby).
```
nodunayo: *Tri-state Problem.

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+
+Although `nil` behaves a lot like `false`, in Ruby they are not quite the same. Having a third state that you can think of as “a bit like `false`” really complicates the logic of your application.
+
+Having `NULL` values in your database means you have to consider how they should behave and also complicates your scopes.
```
nodunayo: By this point, It's like you've spend three paragraphs saying: this is not recommended/it's really bad/confusing, etc., without getting to why.

I think you can condense them and get more quickly to: you have to handle them since in Ruby `nil` and `false` are not quite the same and scopes get complicated too, etc.

_posts/2018/2018-12-17-always-force-booleans-to-be-true-or-false.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+
+### Why not?
+
+If you’re starting from scratch, just do this.
+
+If you already have this issue — and heaven knows I’ve fallen into the trap — you will have to put together a careful migration strategy.
+
+When changing/adding columns or default values to an existing table many databases will lock the table while they write to every row. This means you’ll have to consider downtime or perhaps multiple smaller migrations to ensure you minimise the impact on your application.
```
nodunayo: Is this last paragraph unique to this issue? I don't think so..right? Therefore, not sure if it's relevant.

Maybe you need to expand on the 'careful migration strategy' in the previous paragraph. Also, this isn't really a "Why not?". It's just saying: "Be careful." Is there a point where the migration can become too complex that it's not worth doing?

nodunayo (dismissed):

PR #38, opened 2019-02-03
Assigning has_one loses existing association

andycroll:


_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+layout: article
+category: ruby
+image:
+  base: '2019/be-careful-assigning-to-has-one-relations'
+  alt: 'Broken net'
+  source: "https://unsplash.com/photos/IV--3UEiHlI"
+  credit: "Ben Hershey"
+---
+
+For much of the time, when building relationships between models you typically use `has_many` and `belongs_to`. There are some circumstances where a `has_one` relationship is more appropriate.
```
nodunayo:
```suggestion
Most of the time, when building relationships between models, you typically use `has_many` and `belongs_to`. There are some circumstances where a `has_one` relationship is more appropriate.
```

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+However, the behaviour of `has_one` has some wrinkles in its behaviour that make it a little trickier to deal with. See the side-effects of the generated `#association=` method [in the Rails documentation](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one).
```
nodunayo:
```suggestion
However, the behaviour of `has_one` has some wrinkles that make it a little trickier to deal with. See the side-effects of the generated `#association=` method [in the Rails documentation](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one).
```

nodunayo: - Is there another word instead of 'wrinkles' now that 'in its behaviour' has been removed? Maybe quirks?

- Maybe: "_Read more_ about the side effects..." instead?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+## When…
+
+…assigning an invalid new object means you lose the relationship with the currently related model.
```
nodunayo: Huh? This is a confusing sentence and do you mean a colon instead of a full-stop? 

It's confusing because this concept of an 'invalid new object' comes out of nowhere. I'm having to stop to think about what it means and what 'losing the relationship...' means as well. 

Maybe you need to say more in the intro about this problem that you're going to introduce.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+person.reload.biography
+=> nil
+```
+
+Oh my.
+
+
+## Use…
+
+…a transaction to wrap the deletion of the existing object. Only delete the existing related object if the new object is valid.
```
nodunayo: You can say this more clearly: Use a transaction to make sure the existing object is deleted only if the new object is valid — or something like that...

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+person.reload.biography
+=> #<Biography id: 1, person_id: 1, text: "Writes emails" ...>
+```
+
+Boom.
+
+
+## But why?
+
+The `has_one` relation has always had this behaviour. In the first example the assignment (in this case via `#biography=`) to the `has_one` attribute, removes the previously joined model at the database level even if the newly assigned model isn't valid.
```
nodunayo: Is this first sentence necessary?

nodunayo: No comma after 'attribute'.

nodunayo: The "...removes the previously joined model at the database level..." part...is there a more straightforward way to say this? Also, do we want to say 'association' or 'object' instead of 'model'?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+When this happened in our app, my coworkers and I were surprised!
```
nodunayo: Again, is this necessary? I get you're trying to add story but it seems slightly out of place — probably because it's rather brief and you haven't delved into any specific issues that it caused. I think you should get rid of this, or expand it a little to explain any pain this bug caused.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+The method created in the second solution, uses the `!` variants of the `create` and `destroy` methods inside a transaction to raise exceptions and rollback both changes if either fails.
```
nodunayo: No comma after 'solution'.

nodunayo: This whole section doesn't really explain 'why?' in the way previous 'why?' sections of this newsletter has. It more just explains how the solution works and hints that it'll save you and your colleagues 'surprise' in the future. Is there a way you can express more of the pain/gotchas involved?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+### In Modern Rails 5.2
+
+It’s trickier to shoot yourself in the foot in Rails after version 5.2 due to the default behaviour of `belongs_to` also including a validation. In the examples I used the `optional: true` argument to make the `Person` class behave like pre-5.2 Rails.
+
+When the framework attempts to set the `person_id` attribute to `nil` in Rails 5.2 a `ActiveRecord::RecordNotSaved` error is raised that prevents the assignation even if the new record is valid.
```
nodunayo: I would move this paragraph to be first in this section. Start it with something like: In Rails 5.2, if you attempt to set the `person_id`....

Add a comma after 'Rails 5.2'. 

_An_  `ActiveRecord::RecordNotSaved`...

*which prevents...

Then the second paragraph (or maybe just a sentence at the end of the first paragraph) can just be: I used `optional:true` to mimic...

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+The (slightly inelegant) solution above would also work in a default Rails 5.2 `belongs_to` relationship.
```
nodunayo: Would it only be 'slightly inelegant' in the context of Rails 5.2, or is it always inelegant? If so, why? And is there a more elegant solution?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+### Why not?
+
+There’s a solid argument that using `has_one` has more weird edges than the more typical `has_many` relationship.
```
nodunayo: *It's been argued that using the `has_one` association leads to many more edge cases than the more typical `has_many` association. Therefore, one approach would be to engineer the logic of your application...

nodunayo (dismissed):
I hope these edits makes sense. I felt rather rambly and not direct.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
andycroll: I think we definitely want to say "model" because it's an Active Record thing.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+  alt: 'Broken net'
+  source: "https://unsplash.com/photos/IV--3UEiHlI"
+  credit: "Ben Hershey"
+---
+
+Most of the time, when building relationships between models, you typically use `has_many` and `belongs_to`. There are some circumstances where a `has_one` relationship is more appropriate.
+
+However, the behaviour of `has_one` has some quirks that make it a little trickier to deal with.
+
+When you assign a new instance of a “contained” model to the `has_one` model the existing “contained” model is removed from the association and _causes a permanent change to be written to the database_. This happens whether the new model is valid or not.
```
nodunayo: Getting there. The use of the word 'contained' isn't so clear. Is there another option or way?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+Read about this side-effect of the generated `#association=` method [in the Rails documentation](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one).
```
nodunayo: *side effect

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Instead of…
+
+…directly assigning a new object to a `has_one` object’s `#association=` method and being surprised by the side effects:
```
nodunayo, line 24: Good explanation!

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+person = Person.create(name: "Andy Croll")
+=> #<Person id: 1, name: "Andy Croll" ...>
+person.create_biography(text: "Writes emails")
+=> #<Biography id: 1, person_id: 1, text: "Writes emails" ...>
+person.biography = Biography.new("Looks handsome")
+=> #<Biography id: 2, person_id: 1, text: "Looks handsome" ...>
+Biography.count
+=> 2
+Biography.first
+=> #<Biography id: 1, person_id: nil, text: "Looks handsome" ...>
```
nodunayo, line 48: The ID changes too? And it becomes the first one? Interesting..

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+But I didn't call save?
```
nodunayo: *save...

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
nodunayo: Oh wait — is it that the first biography is edited but it doesn't ultimately associate with the person since it's invalid?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Boom.
+
+
+## But why?
+
+The first example demonstrates that the assignment (in this case via `#biography=`) to the `has_one` attribute removes the previously associated model and writes to the database.
+
+In Rails prior to 5.2.1 there is a bug that _even if_ the newly assigned model was invalid and would not save the existing record would still be changed.
```
nodunayo, line 95: Comma after 'save'.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+The method I’ve created in the second solution uses the `!` variants of the `create` and `destroy` methods inside a transaction to raise exceptions and rollback both the unassignment of the existing model and assignment of the new one if either fails.
+
+This prevents subtle bugs and the unusual behaviour of implicit writing to the database without calling `#save`.
+
+
+### In Modern Rails (after 5.0)
+
+In Rails 5.0, if you attempt to set the `person_id` attribute to `nil`, as the assignation of the new model will do, an `ActiveRecord::RecordNotSaved` error is raised. This prevents the assignation even if the new record is valid.
+
+The model being removed from the association is no longer valid because `belongs_to` relationships now have an [implied presence validation](https://guides.rubyonrails.org/5_0_release_notes.html#active-record-notable-changes). The change in Rails’ behaviour happened in [this pull request](https://github.com/rails/rails/pull/18937).
```
nodunayo: This sentence isn't that clear. Are you saying when you un-assign before reassigning?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+I used the `optional: true` argument, in my example, to make the `Person` class behave more like pre-5.0 Rails.
```
nodunayo: No commas after 'argument' or 'example'

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+The solution I’ve proposed would also work in this more strict `belongs_to` relationship.
```
nodunayo: Maybe you want to say it also works in newer versions of Rails? That's what you're getting at, right? But also — at this point I'm getting confused: with the stricter version, do you still need the fix?

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Why not?
+
+You can manage the `has_one` relationship in multiple other ways that do not wrap the whole task in a `replace_whatever`-style method such as I’ve suggested.
+
+In fact this [bug fix in Rails 5.2.1](https://github.com/rails/rails/pull/32796) ([commit](https://github.com/rails/rails/commit/c87b3346ca6e1d21a6bccb29ccedf0b95fda7abc)) means the `#whatever=` method does now rollback the changes as my solution would.
```
nodunayo: *performs a rollback of the changes, just as my solution would.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,132 @@
...
+
+You could:
+
+  * Avoid calling the `#whatever=` method in your code and use the clearer `#create_whatever!()` instead
+  * Allow assignment only when the association is empty
+  * Declare your association with `autosave: true/false` to make the behaviour explicit in your code
+
+...but you will need to manage it somehow.
+
+There is a reasonable argument that using `has_one` leads to more edge cases than the more typical `has_many` relationship. One approach would be to engineer the logic of your application to use that relationship, which might make for more predictable behaviour.
```
nodunayo, line 131: I have to say — I get to the end of this article and I'm like: wait — so what am I meant to do? Which Rails version am I on? What is best?

Admittedly, I started reading it last night and finished it this afternoon...so I've had a context gap...but there is a lot here.

nodunayo (dismissed):

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
andycroll: There is. But it's all interrelated.

andycroll: https://github.com/andycroll/andycroll.com/pull/38/commits/c94fb096d0936f2542bc86203bc8af667ad17a20 < worth it I think

nodunayo: Instead, maybe it's after the list, where you say: "...but you have to deal with it somehow." Maybe you point out there: the main thing I want you to take away is: be aware of the peculiarities (don't say it exactly like that, but you catch my drift!)

andycroll: Done.

_posts/2019/2019-02-10-be-careful-assigning-to-has-one-relations.md
```diff
@@ -0,0 +1,136 @@
...
+
+You could:
+
+  * Avoid calling the `#whatever=` method in your code and use the clearer `#create_whatever!()` instead
+  * Allow assignment only when the association is empty
+  * Declare your association with `autosave: true/false` to make the behaviour explicit in your code
+
+…but you will need to manage it somehow.
+
+**The main thing to be aware of is the peculiarities of `has_one` relationships, should you choose to use one.**
```
nodunayo: *are

PR #40, opened 2019-03-06
Don’t use instance variables in partials

andycroll:


_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+layout: article
+category: ruby
+image:
+  base: '2019/dont-use-instance-variables-in-partials'
+  alt: 'Partial eclipse'
+  source: "https://unsplash.com/photos/NbgQfUvKFE0"
+  credit: "Mark Tegethoff"
+---
+
+Rails’ view architecture is a flexible and magical place. Perhaps too much.
```
nodunayo: *Perhaps too much _so_.

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+Instance variables from controllers, like `@whatever`, are magically available in view templates. Inferred template names. A global namespace of helpers. For some it's too much, but most of us accept the “magic” and muddle through any strangeness.
```
nodunayo: Do we need the 'like @whatever' part? Maybe just say "Instance variables defined in controller files"?

nodunayo: > Inferred template names. A global namespace of helpers.

This doesn't make sense to me. You need to be more explicit...or word it differently.

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+## Instead of…
+
+…using the global variables from your template inside your partials, pass the values as local variables.
```
nodunayo: I think remove the '...pass the values...' part of this sentence, as I got confused thinking the below was the solution.

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+</div>
+...
+```
+
+
+## But why?
+
+It’s too easy to presume that certain instance variables, in this case `@album`, are available in the context your partial is called in.
+
+When you first extract the partial, for organisation, it is only used in one place. When you reuse the partial somewhere else in your application the controller action in which it is eventually called must also correctly assign that exact variable.
```
nodunayo: *When you first extract a partial, often for organisational reasons, it is typically used in only one place.

nodunayo: * comma after 'somewhere else in your application'

nodunayo: * "correctly assign the same variable from the original controller action"

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+As your application becomes more complex a view might contain several partials—each expecting their own instance variables—so you’ll need to assign several instance variables in the controller action. The context of these assignments will be a long way from the nested partial in which they’re used.
```
nodunayo: * comma after 'more complex'
* long way _away_

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+By not using instance variables your partials also benefit from simpler reuse.
+
+This “early magic, later complexity” is one of the issues that people have with Rails’ HTML rendering environment.
```
nodunayo: What you're trying to say here is issues come about when people rely on magic early on, and then they're dealing with a more complex application later?

I think you should be more explicit in this paragraph about that.

nodunayo: Then after that, I'd add the: if you avoid using instance variables in your partials, they become simpler to reuse, even as your application grows in complexity (essentially moving what's currently line 83).

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+### A little help
+
+You can find lines in view partials which reference an instance variable by using this command from the root of your application in your Terminal.
```
nodunayo: Colon instead of full stop?

nodunayo: Should 'Terminal' be capitalised?

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+```shell
+grep -rEHn "\@\w+" --include "_*.erb" app/views
+```
+
+
+
+## Why not?
+
+This is generally good practice. It might feel easier to leave any instance variables in place and also can feel redundant to write the additional code to define the partial’s variables in the call to `render`. But it will bite you in the end.
```
nodunayo:
```suggestion
It might feel easier to leave any instance variables in place and it can also feel redundant to write the additional code needed in order to define the partial’s variables in the call to `render`. However, not making these changes will bite you in the end.
```

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+It's certainly _never_ a good idea to create an instance variable to share it between partials.
```
nodunayo, line 104: *for the sole purpose of sharing it between partials

_posts/2019/2019-03-10-dont-use-instance-variables-in-partials.md
```diff
@@ -0,0 +1,104 @@
...
+
+Note that there are other issues in the example I've used. It looks like the call to `character#features_on_tracks` will probably invoke some sort of database query.
```
nodunayo: Is this needed?

nodunayo (dismissed):

andycroll:
I'm gonna merge this unless you have any major objections. I realise I was rubbish and late this week!

PR #45, opened 2019-04-19
Where.not and nil

andycroll:


_posts/2019/2019-03-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,45 @@
...
+layout: article
+category: ruby
+image:
+  base: '2019/be-aware-of-nil-values-when-using-where-not'
+  alt: 'Half-finished painted wall'
+  source: "https://unsplash.com/photos/oBb-Y26PJgg"
+  credit: "Samuel Zeller"
+---
+
+The use of `where.not` when building Active Record scopes can help you to elegantly specify scopes where an attribute does not have a specific value.
```
nodunayo: *for cases where an attribute... ?

_posts/2019/2019-03-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,45 @@
...
+
+However the abstraction has some quirks if you have fields that can be `NULL` in your database (`nil` in Ruby).
```
nodunayo: *...if that attribute can be `NULL` — is that what you're trying to say?

_posts/2019/2019-03-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,45 @@
...
+…expecting `where.not` to return `nil` values:
+
+```ruby
+non_soy_options = Coffee.where.not(milk: "Soy")
+```
+
+
+## Use…
+
+…an `or` clause to explicitly request the `nil` values.
```
nodunayo: Oh! So I think you want to say at the beginning maybe more explicitly that it doesn't return 'nil' columns by default and there are cases where you may want/expect that.

_posts/2019/2019-03-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,45 @@
...
+
+```ruby
+non_soy_options = Coffee.where.not(milk: "Soy").or(Coffee.where(milk: nil))
+```
+
+
+## But why?
+
+Of course this all depends on whether the answer you're looking for is expected to include nil values!
```
nodunayo: *whether you're expecting/wanting/needing the answer to include nil values — something like that.

_posts/2019/2019-03-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,45 @@
...
+
+This is because in SQL the results of `WHERE milk = "Soy"` aren’t the inverse of `WHERE milk != "Soy"`. Neither of them include the case `WHERE milk IS NULL`
```
nodunayo: But obviously `WHERE milk = "Soy"` wouldn't include cases where milk is NULL...so the 'neither of them' statement doesn't make sense to me.

And to confirm: `WHERE milk != "Soy"` will also not include cases where milk is NULL, right?

nodunayo (dismissed):

_posts/2019/2019-04-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,47 @@
...
+image:
+  base: '2019/be-aware-of-nil-values-when-using-where-not'
+  alt: 'Half-finished painted wall'
+  source: "https://unsplash.com/photos/oBb-Y26PJgg"
+  credit: "Samuel Zeller"
+---
+
+The use of `where.not` when building Active Record scopes can help you to elegantly specify scopes for cases where an attribute does not have a specific value.
+
+However the abstraction has some quirks if the attribute you’re querying can be `NULL` in your database (`nil` in Ruby). The `where.not` scope doesn’t automatically return `nil` values unless you specify them.
```
nodunayo: *unless you specify that you want them (i.e. want them returned)?

_posts/2019/2019-04-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,47 @@
...
+```ruby
+non_oat_options = Coffee.where.not(milk: "Oatly").or(Coffee.where(milk: nil))
+```
+
+
+## But why?
+
+Of course this all depends on whether you are expecting the answer to include those nil values!
+
+The standard usage `where(milk: "Oatly")`, as expected, does not include `nil` values. However in Rubyit looks as though `where.not(milk: "Oatly")` would be the inverse: _but it isn’t_.
```
nodunayo: *Ruby it

nodunayo:
```suggestion
`where(milk: "Oatly")`, as expected, will not return `nil` values, and you would think that `where.not(milk: "Oatly")` would be the complete inverse: _but it isn’t_.
```

Not sure if 'complete' makes sense here, but it does to me...to get across: 'everything else' not returned in `where(milk: "Oatly")`.

_posts/2019/2019-04-28-be-aware-of-nil-values-when-using-where-not.md
```diff
@@ -0,0 +1,47 @@
...
+
+This is because the SQL result of a `where.not` scope is SQL’s `!=` operator which doesn't return `NULL` values from the database.
```
nodunayo: * comma before 'which'.

nodunayo (dismissed):

PR #117, opened 2021-02-07
use url helpers

andycroll:


_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+layout: article
+category: ruby
+image:
+  base: "2021/url-helpers-outside-views-controllers"
+  alt: "Abstract greyscale roads"
+  source: "https://unsplash.com/photos/36b7JBzhfF4"
+  credit: "Bogdan Karlenko"
+---
+
+Occasionally you'll need to generate a URL for your own application outside of the normal places; views or controllers.
```
nodunayo: Do you mean outside of "the normal places, views, or controllers"...or "outside of the normal places, which are views or controllers"?

Either way, I'd update the way the list is written, and drop the semicolon.

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+      "http://userinfoapi.com/",
+      body: {callback_to: "https://myapp.com/user/#{user.id}"})
+  end
+end
+```
+
+
+## Include...
+
+...the route helpers that Rails includes automatically in controllers.
```
nodunayo, line 33: Use rather than "Include" since you use "includes" a few words later.

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+  def perform(user)
+    Net::HTTP.post(
+      "http://userinfoapi.com/",
+      body: {callback_to: user_url(user, host: "myapp.com")})
+  end
+end
+```
+
+You'll need to ensure you specify a `host` for any URLs you generate as the URL helpers are expecting the context of an HTTP request (which is always available in the views and controllers) which sets that option.
```
nodunayo: This is confusing. There is definitely a simpler way to say this!

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+An enhanced version of this pattern is to use a concern and piggyback on the (probably) already set Action Mailer url options.
```
nodunayo: *an ActiveSupport concern (just to be a bit clearer)

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+      "http://userinfoapi.com/",
+      body: {callback_to: user_url(user)})
+  end
+enda
+```
+
+
+## Why?
+
+It's better to use the URL helpers throughout your application, they're convenient and consistent. Should you change a route for any reason and fail to change the URL method, your tests will alert you to the problem. This breakage wouldn't happen if you interpolate an `id` into a hard-coded URL.
```
nodunayo:
```suggestion
It's better to use the URL helpers throughout your application since they're convenient and consistent. Should you change a route for any reason and fail to update the relevant URL method, your tests will alert you to the problem. You wouldn't get this advance warning if you had interpolated an `id` into a hard-coded URL.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+As demonstrated, I typically find I need this pattern when calling out to external APIs that require a webhook to listen for a response.
```
nodunayo:
```suggestion
As demonstrated in the example above, I typically find I need this pattern when calling out to external APIs that require a webhook to listen for a response.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+Another use case might be if you're writing an API for your application that includes a field describing URLs of resources in your application. However in that case you might want to build or use tooling, such as [`active_model_serializers`](https://github.com/rails-api/active_model_serializers), to better support it. API responses are more akin to views and probably deserve their own
```
nodunayo: The first sentence here isn't super clear.

*However, in that case, it might be better to build...

Also, is this paragraph complete?

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+## Why not?
+
+The lack of the automatically included route helpers can be an indication that this might be a place you _shouldn't_ be using your routes. If you're using this technique, make sure you're not over-using it.
```
nodunayo: Again, not clear. I think you just need to be more explicit in this issue.

e.g. in the case: "If you're trying to use the route helpers and find a place in your code where they aren't automatically included, then this might be an indication that..."

And perhaps drop the 'over-using' sentence as there's no real benchmark for people to know what 'over-using' is and your newsletter errs very much on the practical side of things!

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+If you're including this concern in an Active Record model, there's normally a better place to do it.
```
nodunayo: Following on from this, I think remove this (as the route helpers aren't automatically included in the models, right?).

Unless you want to say: definitely avoid using them in an AR model. There's almost certainly a better place!

nodunayo (dismissed):

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
andycroll:
```suggestion
Occasionally you'll need to generate a URL for your own application outside views or controllers where your URL helpers are available. These helper methods are [generated form your routes](https://guides.rubyonrails.org/routing.html#path-and-url-helpers), e.g. `user_books_path(user)`.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
andycroll:
```suggestion
An enhanced version of this pattern is to use an Active Support concern and piggyback on the (probably) already set Action Mailer url options.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
andycroll: You might also use this if you're writing an API for your application that exports a field describing URLs of resources in your application.

```json
{
  "id": 23432,
  "name": "Nadia",
  "links": {
    "self": "https://yourapp.com/user/23432.json"
  }
}
```

However in this case you might want to use tooling, such as [`active_model_serializers`](https://github.com/rails-api/active_model_serializers), to better support it. API responses are more akin to views and probably deserve their own dedicated approach.

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
andycroll:
```suggestion
If you're trying to use route helpers in a place where they aren't already included, this might be an indication that this is a place you _shouldn't_ be using them.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+  def perform(user)
+    Net::HTTP.post(
+      "http://userinfoapi.com/",
+      body: {callback_to: "https://myapp.com/user/#{user.id}"})
+  end
+end
+```
+
+
+## Include...
```
andycroll:
```suggestion
## Use...
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+As demonstrated in the example above, I typically find I need this pattern when calling out to external APIs that require a webhook to listen for a response.
+
+Another use case might be if you're writing an API for your application that includes a field describing URLs of resources in your application. However in that case you might want to build or use tooling, such as [`active_model_serializers`](https://github.com/rails-api/active_model_serializers), to better support it. API responses are more akin to views and probably deserve their own
+
+
+## Why not?
+
+If you're trying to use route helpers in a place where they aren't already included, this might be an indication that this is a place you _shouldn't_ be using them.
+
+If you're including this concern in an Active Record model, there's normally a better place to do it.
```
andycroll:
```suggestion
Avoid using your routes within an Active Record model, there's almost certainly a better place!
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
andycroll:
```suggestion
URL helpers are normally used in the context of a web request. The view or controller action will be provided the `host` or domain by the application. Outside of this context you'll need to ensure you specify a `host` for any helpers ending in `_url`.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+layout: article
+category: ruby
+image:
+  base: "2021/url-helpers-outside-views-controllers"
+  alt: "Abstract greyscale roads"
+  source: "https://unsplash.com/photos/36b7JBzhfF4"
+  credit: "Bogdan Karlenko"
+---
+
+Occasionally you'll need to generate a URL for your own application outside views or controllers where your URL helpers are available. These helper methods are [generated form your routes](https://guides.rubyonrails.org/routing.html#path-and-url-helpers), e.g. `user_books_path(user)`.
```
nodunayo:
```suggestion
Occasionally you'll need to generate a URL to your own application, outside of views or controllers. In these places, URL helpers are available. These helper methods are [generated from your routes](https://guides.rubyonrails.org/routing.html#path-and-url-helpers), e.g. `user_books_path(user)`.
```

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
```diff
@@ -0,0 +1,89 @@
...
+
+  def perform(user)
+    Net::HTTP.post(
+      "http://userinfoapi.com/",
+      body: {callback_to: user_url(user, host: "myapp.com")})
+  end
+end
+```
+
+URL helpers are normally used in the context of a web request. The view or controller action will be provided the `host` or domain by the application. Outside of this context you'll need to ensure you specify a `host` for any helpers ending in `_url`.
```
nodunayo:
```suggestion
URL helpers are normally used in the context of a web request. In these cases, the view or controller action is given the `host` or domain of the request by the application. Outside of this context you'll need to ensure you specify a `host` for any helpers ending in `_url`.
```

nodunayo: I still don't think this paragraph is super clear or tight.

_posts/2021/2021-02-22-url-helpers-outside-views-controllers.md
nodunayo: Hmmmm. Maybe it's not clear because I'm not seeing how the url helpers would be used here. Do you mean for the `self` property, you'd use them? But instead, you're saying there's probably a better way.

I guess it's not clear also why using dedicated tooling would be better here.

Ahh, it's just clicked while writing this comment: an API response is its own thing, so rather than using the URL helpers, use the stuff made to generate API responses, e.g. active_model_serializers.

nodunayo (dismissed):

PR #231, opened 2026-01-08
Add post about Array#rfind in Ruby 4.0

andycroll:
New One Ruby Thing post covering Array#rfind added in Ruby 4.0 for finding the last matching element in an array.


_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
```diff
@@ -0,0 +1,65 @@
...
+category: ruby
+date: 2026-01-12 09:00
+image:
+  base: "2026/find-the-last-matching-element-with-rfind"
+  alt: "Magnifying glass on white table"
+  credit: "Markus Winkler"
+  source: "https://unsplash.com/photos/magnifying-glass-on-white-table-afW1hht0NSs"
+---
+
+Ruby 4.0 landed at Christmas 2025 with a bunch of new features. One [small but useful addition](https://bugs.ruby-lang.org/issues/21678) is `Array#rfind`, which lets you find the last element in an array that matches a condition.
```
nodunayo: *'during' or 'on' instead of 'at'?

* I'd also hyperlink `Array#rfind`.

* which finds the last element in an array..

_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+
+## Why?
+
+The `rfind` method iterates backwards from the last element, returning the first element that matches the given condition. It's the **r**everse of `find`, which starts from the beginning.
+
+This is more efficient than `reverse.find` or `reverse_each.find` because it doesn't create an intermediate reversed array or enumerator in memory. For large arrays, this _can_ make a noticeable difference.
+
+You might wonder why this was added to `Array` specifically rather than `Enumerable`. The `Enumerable` module relies on the `#each` method, which only works in the forward direction. The only way to scan backwards generically would be to convert to an array first, defeating the purpose. Arrays, however, can be traversed in either direction efficiently by their nature.
```
nodunayo: * However, arrays can be...

_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
```diff
@@ -0,0 +1,65 @@
...
+## Why not?
+
+If you're not yet on Ruby 4.0, you'll need to stick with `reverse_each.find` for now. It's not a dramatic difference, but `rfind` is cleaner and more intentional.
+
+For very small arrays, the performance difference is negligible, so the main benefit is readability—your intent is clearer when you use `rfind`.
+
+
+## Did you know?
+
+In the same changeset, [Kevin Newton](https://kddnewton.com) also added a specific implmentation of `Array#find` (and its alias `#detect`) rather than relying on the `Enumerable` version. This is a performance improvement for arrays.
```
nodunayo: *implementation (sp)

nodunayo (dismissed):

_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
andycroll:
```suggestion
Ruby 4.0 landed during Christmas 2025 with a bunch of new features. One [small but useful addition](https://bugs.ruby-lang.org/issues/21678) is [`Array#rfind`](https://docs.ruby-lang.org/en/4.0/Array.html#method-i-rfind), which finds the last element in an array that matches a condition.
```

_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
andycroll:
```suggestion
You might wonder why this was added to `Array` specifically rather than `Enumerable`. The `Enumerable` module relies on the `#each` method, which only works in the forward direction. The only way to scan backwards generically would be to convert to an array first, defeating the purpose. However, arrays can be traversed in either direction efficiently by the nature of their implementation in the Ruby VM.
```

_posts/2026/2026-01-12-find-the-last-matching-element-with-rfind.md
andycroll:
```suggestion
In the same changeset, [Kevin Newton](https://kddnewton.com) also added a specific implementation of `Array#find` (and its alias `#detect`) rather than relying on the `Enumerable` version. This is a performance improvement for arrays.
```

PR #243, opened 2026-02-09
Improve clarity of StringInquirer post

andycroll:


andycroll:
Wildly simplified and rewrote for clarity.

